PR #9, opened 2016-04-29
Rearrange GLM class

titipata:
- custom `GLM` class so that we have more `scikit-learn` flavor on how to train and predict from the data. - update example notebook file


titipata:
This has conflict and too many changes apply. Therefore, I will close this pull-request for now.

PR #27, opened 2016-05-01
Allow output array to be more flexible

titipata:
Now, we accept input of flatten array or column array


titipata:
This should partially solve issue #27

PR #115, opened 2016-05-24
Main Makefile to automate tasks

marquesVF:
A proposal to solve the issue #100.
See inclusion in documentation.


Makefile
```diff
@@ -0,0 +1,71 @@
...
+
+# Test related commands
+NOSETESTS ?= nosetests
+
+.PHONY: test-all
+test-all:
+	nosetests
+
+# Examples related commands
+.PHONY: run-poisson
```
jasmainak: do we really need this?

Makefile
```diff
@@ -0,0 +1,71 @@
...
+run-poisson:
+	python examples/api/plot_poisson.py
+
+.PHONY: run-poisson2
```
jasmainak: same question here

jasmainak:
we should also have a `make clean`, `make test` and `make all` option.

jasmainak:
I see some merge commits. Can you remove them? thanks @marquesVF for taking a stab at this!

Makefile
marquesVF: I see your point. These packages aren't necessary to make Pyglmnet works. They are useful to run the examples only. I gonna remove them and the test commands.

marquesVF: Matplotlib package is needed to build the documentation. Thus, I suggest to add a command line to install it on the `doc-dependencies` target.

marquesVF:
I'm not sure what supposed to be in those targets: `make clean`, `make test` and `make all`. I suppose that `make all` would install the dependencies and the package itself, install the dependencies to build the documentation and build it.

jasmainak:
@marquesVF here is what they would do:

`make clean`: Removes `*.pyc` files etc. This is sometimes required when you add/remove files from your package and you don't want the old `*.pyc` files to create import errors
`make test`: Runs the test for the code and also the doc-tests.
`make all`: It would run `make clean`, `make test` and `make doc`

Did you check the code I had linked in #100 ? You can basically copy much of the stuff from there. We shouldn't try to innovate too much here. Just do what others already do well.

Makefile
jasmainak: On Wed, May 25, 2016 at 6:00 PM, Vinicius Marques <notifications@github.com>
wrote:

> In Makefile
> <https://github.com/pavanramkumar/pyglmnet/pull/115#discussion_r64598704>:
>
> > +	pip install nose
> > +	@echo "#### All dependencies were installed successfully ####"
> > +
> > +# Test related commands
> > +NOSETESTS ?= nosetests
> > +
> > +.PHONY: test-all
> > +test-all:
> > +	nosetests
> > +
> > +# Examples related commands
> > +.PHONY: run-poisson
> > +run-poisson:
> > +	python examples/api/plot_poisson.py
> > +
> > +.PHONY: run-poisson2
>
> Matplotlib package is needed to build the documentation. Thus, I suggest
> to add a command line to install it on the doc-dependencies target.
>

sounds good to me.

marquesVF:
I've made the changes and solve the problem of multiples merge commits at #117. I think this PR can be closed.

jasmainak:
@marquesVF you can use `git rebase` to get rid of merge commits and then force push so you avoid having to create another PR :)

PR #123, opened 2016-06-17
Improve gradients

pavanramkumar:
Earlier, the loss and gradients were computed as the sum of likelihood  terms over samples. This created a dependency of gradient magnitude and convergence rate on `n_samples`. I've fixed these now to be more similar to `sklearn`


pavanramkumar:
I'm still not satisfied with it. Three parameters still need to be balanced: `learning_rate`', `max_iter`, and `tol`. 

But the core algorithms work and they can now be benchmarked against `sklearn` with the same values of `lambda` resulting in the same level of regularization in both `pyglmnet` and `sklearn`

PR #130, opened 2016-07-14
Pure python implementation of Newton coordinate descent solver

pavanramkumar:
The multinomial gradient and hessian is currently not yet implemented, but I have tested that it works for all other `distr`. 

My `%timeit` benchmarks suggest that it is 3x-10x slower than 'batch-gradient'.

@jasmainak @raghavrv look at the new methods:
`_cdfast()`, `_gradhess_logloss_1d()`

The goal is to translate these to `cython` or wrap them correctly with `numba`.

I have also added a `cheatsheet.pdf` in the `latex` folder. Have a look there in case of bugs in my code.

Enjoy your liberty and equality!


pavanramkumar:
The new tests suggest that nothing breaks with the `cdfast` solver. I have also verified with all distr that the results are comparable to the `batch-gradient` solver. 

I will merge this PR and then start hacking the cython version in a new PR.

PR #131, opened 2016-07-18
[DOC] pyglmnet application and comparison against scikit-learn - REFACTORING

marquesVF:
This new PR is an approach to solve the problems discussed at #122 regarding the pyglmnet application. I will make another one for the developer documentation.


pavanramkumar:
I believe this PR is now obsolete

PR #150, opened 2016-09-20
[MRG] tikhonov regularization

pavanramkumar:
- implemented Tikhonov both for solvers (`batch-gradient` and `cdfast`)
- tested that it runs better than ridge regression on simulated data
- unit test

todo:
- API example
- application example


pavanramkumar:
Merging this now so that I can work on `pipelines` compatibility. I will do examples in a separate PR along with group lasso. Closes #139

PR #151, opened 2016-09-21
[MRG] Refactored score function, tests and examples to be pipeline compatible

pavanramkumar:
- The `GLM` class accepts an attribute `score_metric` to decide whether to return `'deviance'` or `'pseudo_R2'`. Default is `'deviance'`
- After `fit()` the mean of the training set is stored in `self.ynull_`. This is used to compute 'pseudo_R2'.
- If a list of estimators calls `score()` it returns a list of floats, one for each `reg_lambda`.
- Tests verify that `score()` cannot be called unless `fit()` is run.
- The use of `score()` has been changed in all appropriate examples.


pavanramkumar:
This addresses issues #58 and #98 and is in lieu of PRs #124 and #126

PR #168, opened 2016-10-20
update links after pavanramkumar --> glm-tools transfer

pavanramkumar:


pavanramkumar:
related to #166 circle CI integration still not working because circle CI doesn't recognize the new organization.

PR #172, opened 2016-11-08
updated crime example with dataset fetcher

pavanramkumar:
- wrote a dataset fetcher to grab `communities.data`
- removed missing values
- removed comparison with `scikit-learn` (we should use `BenchmarkGLM()` for all comparisons)

this addresses #149

### DIFF
--- a/examples/plot_community_crime.py
+++ b/examples/plot_community_crime.py
@@ -1,20 +1,17 @@
 # -*- coding: utf-8 -*-
 """
-============================================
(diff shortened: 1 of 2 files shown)


pavanramkumar:
@marquesVF i was able to get to this task: https://github.com/glm-tools/pyglmnet/issues/149#issuecomment-256981017 in this PR; just FYI

pavanramkumar:
closes #149

PR #270, opened 2018-09-13
[MRG+1] Fix issue #269.

geektoni:
Changes summary

-  Update `simulate_glm` to use a `RandomState` instance initialized with `random_state`;
-  Update `GLM` to use a `RandomState` instance initialized with `random_state`;
-  Add a test to ensure correctness of the results when setting `random_state` (this may require a more close review);
- Update `what's_new.rst` with the bugfix news.

### DIFF
--- a/pyglmnet/pyglmnet.py
+++ b/pyglmnet/pyglmnet.py
@@ -352,14 +352,13 @@ def simulate_glm(distr, beta0, beta, X, eta=2.0, random_state=None,
     if not sample:
         return _lmb(distr, beta0, beta, X, eta)
 
-    if random_state is not None:
-        np.random.RandomState(random_state)
+    _random_state = np.random.RandomState(random_state)
     if distr == 'softplus' or distr == 'poisson':
-        y = np.random.poisson(_lmb(distr, beta0, beta, X, eta))
+        y = _random_state.poisson(_lmb(distr, beta0, beta, X, eta))
     if distr == 'gaussian':
-        y = np.random.normal(_lmb(distr, beta0, beta, X, eta))
+        y = _random_state.normal(_lmb(distr, beta0, beta, X, eta))
     if distr == 'binomial' or distr == 'probit':
-        y = np.random.binomial(1, _lmb(distr, beta0, beta, X, eta))
+        y = _random_state.binomial(1, _lmb(distr, beta0, beta, X, eta))
     if distr == 'gamma':
         mu = _lmb(distr, beta0, beta, X, eta)
         y = np.exp(mu)
@@ -486,6 +485,7 @@ class GLM(BaseEstimator):
         self.eta = eta
         self.score_metric = score_metric
         self.random_state = random_state
+        self.rng = np.random.RandomState(self.random_state)
         self.callback = callback
         self.verbose = verbose
         set_log_level(verbose)
@@ -675,7 +675,6 @@ class GLM(BaseEstimator):
         self : instance of GLM
             The fitted model.
         """
-        np.random.RandomState(self.random_state)
 
         # checks for group
         if self.group is not None:
@@ -698,9 +697,10 @@ class GLM(BaseEstimator):
         # Initialize parameters
         beta = np.zeros((n_features + 1,))
         if self.beta0_ is None and self.beta_ is None:
-            beta[0] = 1 / (n_features + 1) * np.random.normal(0.0, 1.0, 1)
+            beta[0] = 1 / (n_features + 1) * \
+                self.rng.normal(0.0, 1.0, 1)
             beta[1:] = 1 / (n_features + 1) * \
-                np.random.normal(0.0, 1.0, (n_features, ))
+                self.rng.normal(0.0, 1.0, (n_features, ))
         else:
             beta[0] = self.beta0_
             beta[1:] = self.beta_
--- a/tests/test_pyglmnet.py
+++ b/tests/test_pyglmnet.py
@@ -1,7 +1,7 @@
 from functools import partial
 
 import numpy as np
-from numpy.testing import assert_allclose
+from numpy.testing import assert_allclose, assert_array_equal
 from pytest import raises
 
 import scipy.sparse as sps
@@ -345,3 +345,40 @@ def test_cdfast():
 def test_fetch_datasets():
     """Test fetching datasets."""
     datasets.fetch_community_crime_data('/tmp/glm-tools')
+
+
+def test_random_state_consistency():
+    """ Test model's random_state """
+
+    # Generate the dataset
+    n_samples, n_features = 1000, 10
+
+    beta0 = 1. / (np.float(n_features) + 1.) * np.random.normal(0.0, 1.0)
+    beta = 1. / (np.float(n_features) + 1.) * \
+        np.random.normal(0.0, 1.0, (n_features,))
+    Xtrain = np.random.normal(0.0, 1.0, [n_samples, n_features])
+
+    ytrain = simulate_glm("gaussian", beta0, beta, Xtrain,
+                          sample=False, random_state=42)
+
+    # Test simple glm
+    glm_a = GLM(distr="gaussian", random_state=1)
+    ypred_a = glm_a.fit_predict(Xtrain, ytrain)
+    glm_b = GLM(distr="gaussian", random_state=1)
+    ypred_b = glm_b.fit_predict(Xtrain, ytrain)
+    ypred_c = glm_b.fit_predict(Xtrain, ytrain)
+
+    # Consistency between two different models
+    assert_array_equal(ypred_a, ypred_b)
+    # Consistency between different run of the same model
+    assert_array_equal(ypred_b, ypred_c)
+
+    # Test also cross-validation
+    glm_cv_a = GLMCV(distr="gaussian", cv=3, random_state=1)
+    ypred_a = glm_cv_a.fit_predict(Xtrain, ytrain)
+    glm_cv_b = GLMCV(distr="gaussian", cv=3, random_state=1)
+    ypred_b = glm_cv_b.fit_predict(Xtrain, ytrain)
+    ypred_c = glm_cv_b.fit_predict(Xtrain, ytrain)
+
+    assert_array_equal(ypred_a, ypred_b)
+    assert_array_equal(ypred_b, ypred_c)
--- a/doc/whats_new.rst
+++ b/doc/whats_new.rst
@@ -23,6 +23,7 @@ Changelog
 BUG
 ~~~
 
+    - Fixed incorrect usage of the random_state parameter by `Giovanni De Toni`_.
     - Fixed incorrect proximal operator for group lasso by `Yu Umegaki`_.
     - Changed stopping criteria for convergence (a threshold on the change in objective value) which
       stopped too learly. The new criteria is a threshold on the norm of the gradient, by `Pavan Ramkumar`_. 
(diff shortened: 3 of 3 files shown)


pyglmnet/pyglmnet.py
```diff
@@ -322,7 +322,7 @@ def _gradhess_logloss_1d(distr, xk, y, z, eta):
     return 1. / n_samples * gk, 1. / n_samples * hk
 
 
-def simulate_glm(distr, beta0, beta, X, eta=2.0, random_state=None,
+def simulate_glm(distr, beta0, beta, X, eta=2.0, random_state=0,
```
jasmainak: can you keep it `None`? It should not be deterministic by default

tests/test_pyglmnet.py
```diff
@@ -345,3 +345,42 @@ def test_cdfast():
...
     """Test fetching datasets."""
     datasets.fetch_community_crime_data('/tmp/glm-tools')
+
+def test_random_state_consistency():
+    """ Test model's random_state """
+
+    # Generate the dataset
+    n_samples, n_features = 1000, 100
+
+    np.random.seed(42)
```
jasmainak: don't set global seed ... it's not good practice

tests/test_pyglmnet.py
```diff
@@ -345,3 +345,42 @@ def test_cdfast():
...
+
+            ytrain = simulate_glm(distr, beta0, beta, Xtrain,
+                                   sample=False, random_state=42)
+
+            glm = GLM(distr=distr, solver=solver, alpha=0.01, reg_lambda=0.1, random_state=1)
+            glm.fit(Xtrain,ytrain)
+            result_a = glm.predict(Xtrain)
+
+            # Train again the same model
+            glm = GLM(distr=distr, solver=solver, alpha=0.01, reg_lambda=0.1, random_state=1)
```
jasmainak: can you edit an existing test rather than adding a completely new test?

tests/test_pyglmnet.py
```diff
@@ -345,3 +345,42 @@ def test_cdfast():
...
+            glm.fit(Xtrain, ytrain)
+            result_b = glm.predict(Xtrain)
+
+            print(result_a)
```
jasmainak: usually you don't put `print` statements in tests

pyglmnet/pyglmnet.py
```diff
@@ -486,6 +485,7 @@ def __init__(self, distr='poisson', alpha=0.5,
         self.eta = eta
         self.score_metric = score_metric
         self.random_state = random_state
+        self.random_generator = np.random.RandomState(self.random_state)
```
jasmainak: I would call it `rng` instead of `random_generator` as in `sklearn`

tests/test_pyglmnet.py
```diff
@@ -345,3 +345,42 @@ def test_cdfast():
...
+
+            for i in range(len(result_a)):
+                assert(result_a[i] == result_b[i])
```
jasmainak: missing empty line at the end of file ... we follow pep8 conventions

pyglmnet/pyglmnet.py
```diff
@@ -352,14 +352,13 @@ def simulate_glm(distr, beta0, beta, X, eta=2.0, random_state=None,
     if not sample:
         return _lmb(distr, beta0, beta, X, eta)
 
-    if random_state is not None:
-        np.random.RandomState(random_state)
+    _random_state = np.random.RandomState(random_state)
```
jasmainak: see comment above. `random_state` should be `None` by default

jasmainak:
@geektoni make sure to update `whats_new.rst` once you have addressed all the comments

tests/test_pyglmnet.py
geektoni: I would be slightly more in favour to keep this new test since it avoids to have only one big test that checks for all the possible cases (I would be happy to do a refactoring of tests in order to reduce copy-paste code). However, if you think it would be better to move this test into another one, I will update it accordingly.

pyglmnet/pyglmnet.py
geektoni: What do you mean here? `random_state` has as default `None`, therefore the `RandomState` object won't give deterministic result so the original behaviour should be preserved.

jasmainak: okay I thought the numpy `RandomState` container does not accept `None`. But it's good. Looks like it does ...

tests/test_pyglmnet.py
jasmainak: I see what you mean. But I think it's overkill to test all the solvers and distributions for this. It will increase the testing time. We just need to run a `fit_predict` twice on a GLM object (and maybe also GLMCV?) and compare their results. So, it shouldn't be more than 3-4 lines of code

geektoni: I updated the test to run only with `fit_predict`, without checking for all the possible solvers/distributions combinations. You are also right about `GLMCV`, so I've added it as well.
fortunately, this test case does not take so long to run in the end.

geektoni:
I've addressed the last comments and updated the `whats_new` file.

tests/test_pyglmnet.py
```diff
@@ -345,3 +345,35 @@ def test_cdfast():
...
+    ytrain = simulate_glm("gaussian", beta0, beta, Xtrain,
+        sample=False, random_state=42)
+
+    # Test simple glm
+    glm = GLM(distr="gaussian", random_state=1)
+    result_a = glm.fit_predict(Xtrain,ytrain)
+    result_b = glm.fit_predict(Xtrain,ytrain)
+
+    for i in range(len(result_a)):
+        assert(result_a[i] == result_b[i])
```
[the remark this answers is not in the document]
geektoni: mmh yeah, that's surely better.

tests/test_pyglmnet.py
```diff
@@ -345,3 +345,35 @@ def test_cdfast():
...
+    Xtrain = np.random.normal(0.0, 1.0, [n_samples, n_features])
+
+
+    ytrain = simulate_glm("gaussian", beta0, beta, Xtrain,
+        sample=False, random_state=42)
+
+    # Test simple glm
+    glm = GLM(distr="gaussian", random_state=1)
+    result_a = glm.fit_predict(Xtrain,ytrain)
+    result_b = glm.fit_predict(Xtrain,ytrain)
```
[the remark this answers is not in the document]
geektoni: Given that I have created a GLM model and I've set its `random_state` to a value != None, I wanted to check that if I were to call multiple time `fit_predict` (or run multiple time the script) I would always get the same results. Before this patch, this was not the case.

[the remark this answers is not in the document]
pavanramkumar: I would do it like so:
```
glm_a = GLM(distr='gaussian', random_state=1)
result_a = glm_a.fit_predict(Xtrain, ytrain)
glm_b = GLM(distr='gaussian', random_state=1)
result_b = glm_b.fit_predict(Xtrain, ytrain)

File: tests/test_pyglmnet.py [hunk 10]
```diff
@@ -345,3 +345,35 @@ def test_cdfast():
...
+
+    beta0 = 1. / (np.float(n_features) + 1.) * \
+            np.random.normal(0.0, 1.0)
+    beta = 1. / (np.float(n_features) + 1.) * \
+           np.random.normal(0.0, 1.0, (n_features,))
+    Xtrain = np.random.normal(0.0, 1.0, [n_samples, n_features])
+
+
+    ytrain = simulate_glm("gaussian", beta0, beta, Xtrain,
+        sample=False, random_state=42)
```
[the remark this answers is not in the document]
geektoni: Sure! I am using `flake8 --count pyglmnet` to check the code style (as suggested in the developing guide) but apparently it is not enough :/

tests/test_pyglmnet.py
```diff
@@ -345,3 +345,35 @@ def test_cdfast():
...
+    result_a = glm.fit_predict(Xtrain,ytrain)
+    result_b = glm.fit_predict(Xtrain,ytrain)
+
+    for i in range(len(result_a)):
+        assert(result_a[i] == result_b[i])
+
+    # Test also cross-validation
+    glm_cv = GLMCV(distr="gaussian", cv=3, random_state=1)
+    result_a_cv = glm_cv.fit_predict(Xtrain, ytrain)
+    result_b_cv = glm_cv.fit_predict(Xtrain, ytrain)
```
pavanramkumar: same comment as above:
```
glm_cv_a = GLMCV(distr="gaussian", cv=3, random_state=1)
result_a_cv = glm_cv_a.fit_predict(Xtrain, ytrain)
glm_cv_b = GLMCV(distr="gaussian", cv=3, random_state=1)
result_b_cv = glm_cv_b.fit_predict(Xtrain, ytrain)
```

tests/test_pyglmnet.py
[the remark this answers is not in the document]
jasmainak: I think both are worth checking. Same GLM object but doing `fit_predict` twice should also give the same result. And different GLM object but same random state should also give same `fit_predict` result ...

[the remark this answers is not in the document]
pavanramkumar: Sure, but given that we update a few hidden instance variables, especially in `GLMCV()`, I think we'll have to do the same object, same state, twice `fit_predict` check more carefully.

tests/test_pyglmnet.py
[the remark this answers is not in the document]
geektoni: Nevermind this comment. Now everything should be PEP8 compliant.

tests/test_pyglmnet.py
```diff
@@ -345,3 +345,40 @@ def test_cdfast():
...
+
+    ytrain = simulate_glm("gaussian", beta0, beta, Xtrain,
+                          sample=False, random_state=42)
+
+    # Test simple glm
+    glm_a = GLM(distr="gaussian", random_state=1)
+    result_a = glm_a.fit_predict(Xtrain, ytrain)
+    glm_b = GLM(distr="gaussian", random_state=1)
+    result_b = glm_b.fit_predict(Xtrain, ytrain)
+    result_c = glm_b.fit_predict(Xtrain, ytrain)
```
jasmainak: same comment as above.

tests/test_pyglmnet.py
```diff
@@ -345,3 +345,40 @@ def test_cdfast():
...
+
+    # Consistency between two different models
+    assert_array_equal(result_a, result_b)
+    # Consistency between different run of the same model
+    assert_array_equal(result_b, result_c)
+
+    # Test also cross-validation
+    glm_cv_a = GLMCV(distr="gaussian", cv=3, random_state=1)
+    result_a_cv = glm_cv_a.fit_predict(Xtrain, ytrain)
```
jasmainak: `ypred_a`. No need to create a new variable ...

tests/test_pyglmnet.py
```diff
@@ -345,3 +345,40 @@ def test_cdfast():
...
+    glm_cv_b = GLMCV(distr="gaussian", cv=3, random_state=1)
+    result_b_cv = glm_cv_b.fit_predict(Xtrain, ytrain)
```
jasmainak: `ypred_b`

tests/test_pyglmnet.py
```diff
@@ -345,3 +345,40 @@ def test_cdfast():
...
+    result_c_cv = glm_cv_b.fit_predict(Xtrain, ytrain)
```
jasmainak: you can also call this `ypred_b` if you put the first assert statement above this line

jasmainak:
Good to go for me. @pavanramkumar please merge if you're happy !

pavanramkumar:
This is a super useful contribution! thanks @geektoni and thanks @jasmainak for thoughtful reviews.

PR #346, opened 2019-10-30
Add check for readme to travis

### LINKED ISSUE #337: add check for readme to travis

yathomasi:
Added check for readme to travis using twine. It closes #337

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,9 +15,13 @@ install:
     - conda create -n testenv --yes pip python=$PYTHON
     - source activate testenv
     - conda install --yes --file requirements.txt
+    - pip install --upgrade twine
     - python setup.py develop
+    - python setup.py sdist bdist_wheel
 
 script:
+    - # Checks whether your distribution’s long description will render correctly on PyPI.
+    - twine check dist/*
     - # Run tests, show verbose output and print the top 10 slowest running tests
     - py.test --cov=pyglmnet tests/ --verbose --durations=10
     - # Run doctests
--- a/README.rst
+++ b/README.rst
@@ -8,8 +8,7 @@ A python implementation of elastic-net regularized generalized linear models
 
 `[Documentation (stable version)]`_ `[Documentation (development version)]`_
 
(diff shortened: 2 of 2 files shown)


.travis.yml
```diff
@@ -15,9 +15,16 @@ install:
     - conda create -n testenv --yes pip python=$PYTHON
     - source activate testenv
     - conda install --yes --file requirements.txt
+    - if [ $PYTHON == 3.5 ];
+      then pip install twine==1.15.0;
+      else conda install --yes twine;
+      fi
```
jasmainak: that's a bit weird. Couldn't we just do:

```sh
pip install --upgrade twine
```
or something like that?

yathomasi:
@jasmainak there was dependency problem with conda install on python 3.5 so I tried to bypass that with  if else where simple pip upgrade was just fine.

README.rst
```diff
@@ -8,9 +8,8 @@ A python implementation of elastic-net regularized generalized linear models
 
 `[Documentation (stable version)]`_ `[Documentation (development version)]`_
 
-.. raw:: html
-
-    <img src="pyglmnet-logo.png" height="400px">
+.. image:: https://raw.githubusercontent.com/glm-tools/pyglmnet/master/pyglmnet-logo.png
+   :height: 400px
```
jasmainak: @pavanramkumar what do we do about this?

The height parameter is unfortunately not respected by github which is why we had to use the raw::html directive

jasmainak: let's maybe use this so we don't have to use the `height` parameter:

pyglmnet

The url would be: https://user-images.githubusercontent.com/15852194/67919367-70482600-fb76-11e9-9b86-891969bd2bee.jpg

pavanramkumar: @jasmainak seems like a reasonable solution to use a smaller image. is that was you proposed?

jasmainak:
okay I pushed a commit. @pavanramkumar feel free to merge once CIs come green.

I'm also updating the wiki to remove the instruction about the ..raw::html line

jasmainak:
Merged, thanks @yathomasi !

PR #380, opened 2020-05-13
[MRG] Show progress when training a GLM object.

geektoni:
Addressing issue #379.

GLM objects can show the current training progress statistics (remaining time, iterations/sec, etc.).
This is true also for GLMCV objects.

* Add `tqdm` (https://github.com/tqdm/tqdm/) as a requirement.

### DIFF
--- a/pyglmnet/pyglmnet.py
+++ b/pyglmnet/pyglmnet.py
@@ -7,7 +7,8 @@ import numpy as np
 from scipy.special import expit
 from scipy.stats import norm
 
-from .utils import logger, set_log_level, _check_params
+from .utils import logger, set_log_level, _check_params, \
+    _verbose_iterable, _tqdm_log
 from .base import BaseEstimator, is_classifier, check_version
 
 from .externals.sklearn.utils import check_random_state, check_array, check_X_y
@@ -809,7 +810,7 @@ class GLM(BaseEstimator):
             else:
                 beta = self.beta_
 
-        logger.info('Lambda: %6.4f' % self.reg_lambda)
+        _tqdm_log('Lambda: %6.4f' % self.reg_lambda)
 
         tol = self.tol
         alpha = self.alpha
@@ -820,8 +821,10 @@ class GLM(BaseEstimator):
             ActiveSet = np.ones_like(beta)
 
         self._convergence = list()
+        training_iterations = _verbose_iterable(range(self.max_iter))
+
         # Iterative updates
-        for t in range(0, self.max_iter):
+        for t in training_iterations:
             self.n_iter_ += 1
             beta_old = beta.copy()
             if self.solver == 'batch-gradient':
@@ -861,7 +864,7 @@ class GLM(BaseEstimator):
             if t > 1 and self._convergence[-1] < tol:
                 msg = ('\tParameter update tolerance. ' +
                        'Converged in {0:d} iterations'.format(t))
-                logger.info(msg)
+                _tqdm_log(msg)
                 break
 
             # Compute and save loss if callbacks are requested
@@ -1285,7 +1288,9 @@ class GLMCV(object):
         np.random.shuffle(idxs)
         cv_splits = np.array_split(idxs, self.cv)
 
-        for idx, rl in enumerate(self.reg_lambda):
+        cv_training_iterations = _verbose_iterable(self.reg_lambda)
+
+        for idx, rl in enumerate(cv_training_iterations):
             glm = GLM(distr=self.distr,
                       alpha=self.alpha,
                       Tau=self.Tau,
@@ -1300,7 +1305,7 @@ class GLMCV(object):
                       fit_intercept=self.fit_intercept,
                       random_state=self.random_state,
                       verbose=self.verbose)
-            logger.info('Lambda: %6.4f' % rl)
+            _tqdm_log('Lambda: %6.4f' % rl)
             glm.reg_lambda = rl
 
             scores_fold = list()
--- a/pyglmnet/utils.py
+++ b/pyglmnet/utils.py
@@ -6,9 +6,17 @@ import numpy as np
 from copy import copy
 import logging
 
-
 logger = logging.getLogger('pyglmnet')
 logger.addHandler(logging.StreamHandler())
+
+try:
+    from tqdm.auto import tqdm
+    has_tqdm = True
+except ImportError:
+    logger.warning(
+        "tqdm library not found. Falling back to non-interactive progress "
+        "visualization.")
+    has_tqdm = False
 
 
 def softmax(w):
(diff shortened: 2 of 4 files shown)


geektoni:
Here a little example about how the user would see it (the screen glitch it is not present in real life and it is an artifact created by the screen recorder).

progress

pyglmnet/pyglmnet.py
```diff
@@ -13,6 +13,8 @@
 from .externals.sklearn.utils import check_random_state, check_array, check_X_y
 from .externals.sklearn.utils.validation import check_is_fitted
 
+from tqdm import tqdm
```
jasmainak, line 16: I would question if we really need to add this dependency. Granted it's a light one but still it's better to avoid when possible.

pyglmnet/pyglmnet.py
```diff
@@ -541,7 +543,8 @@ def __init__(self, distr='poisson', alpha=0.5,
                  learning_rate=2e-1, max_iter=1000,
                  tol=1e-6, eta=2.0, score_metric='deviance',
                  fit_intercept=True,
-                 random_state=0, callback=None, verbose=False):
+                 random_state=0, callback=None, verbose=False,
+                 show_progress=True):
```
jasmainak, line 547: what's the difference between `verbose` and `show_progress`? Can we have `verbose` levels like in joblibs or scikit-learn?

pyglmnet/pyglmnet.py
```diff
@@ -561,6 +564,7 @@ def __init__(self, distr='poisson', alpha=0.5,
         self.random_state = random_state
         self.callback = callback
         self.verbose = verbose
+        self.show_progress = show_progress
```
jasmainak, line 567: be careful adding new attributes and parameters to functions. They are easy to add but hard to remove :)

pyglmnet/pyglmnet.py
geektoni: I agree with this point. I've seen that pyglmnet has not many dependencies and since tqdm is a light one I thought that it wouldn't hurt too much. However, since this is basically a really small feature, I guess we could make all of this optional for the final user (e.g., did you install tqdm? good, then you can have better progress visualization. If you don't have tqdm, then we fall back to the original method.)

pyglmnet/pyglmnet.py
geektoni: personally, I see `verbose` more like a utility to print more information about the status of the system (e.g., from simple status messages to debug messages). `show_progress` refers more specifically to give a better user experience by formatting in a nice way the training progress information.

No hard feelings on this though.

pyglmnet/pyglmnet.py
jasmainak: yes agree with approach, you can have a nested import then. It might slow down import of pyglmnet then. But probably not by much

pyglmnet/utils.py
```diff
@@ -6,6 +6,11 @@
 from copy import copy
 import logging
 
+try:
+    from tqdm import tqdm
+    has_tqdm = True
```
geektoni, line 11: @jasmainak this way if tqdm is not found, we can use the default behaviour instead.

pyglmnet/utils.py
```diff
@@ -130,3 +135,22 @@ def set_log_level(verbose):
             raise ValueError('verbose must be of a valid type')
         verbose = logging_types[verbose]
     logger.setLevel(verbose)
+
+
+def return_verbose_iterable(data):
```
geektoni, line 140: We can basically wrap any list/range object with tqdm with this method. It depends on the logger's level, this means that if the user has set `verbose=True`, then he will see the progress bar. This way we don't need any new state variable (e.g., `show_progress`).

requirements.txt
```diff
@@ -7,3 +7,4 @@ pytest-cov
 coverage
 flake8
 pydocstyle
+tqdm
```
jasmainak, line 10: what version are we supporting?

pyglmnet/utils.py
jasmainak: okay looks reasonable to me @geektoni

pyglmnet/utils.py
jasmainak, line 140:
```suggestion
def verbose_iterable(data):
```

pyglmnet/utils.py
```diff
@@ -130,3 +135,22 @@ def set_log_level(verbose):
...
+    """
+    Wrap an iterable object with tqdm. If tqdm is not available
+    or if we did not set the appropriate log level, then we
+    fall back to the classical method
+    :param data: it can be a list or range() object
+    :param verbose: the verbosity level
+    :return: wrapped object
```
jasmainak, line 147: can you document the docstring in a more standard way? You know how it should be :)

jasmainak:
@geektoni need to update the `whats_new` page

pyglmnet/utils.py
```diff
@@ -130,3 +135,22 @@ def set_log_level(verbose):
...
+    """
+    wrapped_data = data
+    if logger.getEffectiveLevel() == logging.INFO:
+        if not has_tqdm:
+            logging.info(
+                "tqdm library not found. Falling back to non-interactive.")
```
jasmainak, line 153: would this send out a message for every reg_lambda in the regularization path? We don't want to be too verbose either.

pyglmnet/utils.py
```diff
@@ -130,3 +135,22 @@ def set_log_level(verbose):
...
+        else:
+            wrapped_data = tqdm(data)
```
jasmainak, line 155: does this work also on Jupyter notebook? have you checked? I remember having issues with tqdm on the notebook.

requirements.txt
geektoni: I set it to the last stable (4.46).

pyglmnet/utils.py
geektoni: Updated. I will need to change my PyCharm autocompletion to match this one :)

pyglmnet/utils.py
geektoni: Yes, it works nicely (see https://pypi.org/project/tqdm/#ipython-jupyter-integration). Honestly, when using cross-validation it can become a bit messy (especially if we throw too many `not converged` messages).

pyglmnet/utils.py
geektoni: Indeed this happens. I'll try to come up with something less annoying.

jasmainak: sounds good!

geektoni: It should work better now.

pyglmnet/utils.py
```diff
@@ -130,3 +137,26 @@ def set_log_level(verbose):
             raise ValueError('verbose must be of a valid type')
         verbose = logging_types[verbose]
     logger.setLevel(verbose)
+
+
+def verbose_iterable(data):
```
jasmainak, line 142: private function? we don't want to expose this in the public API, do we?

jasmainak:
Otherwise, LGTM. Can you set the PR to MRG if you're ready?

pyglmnet/utils.py
geektoni: mmh how can I make this thingy private? I remember you can only do that for class methods. Anyhow, I changed the method's signature to `_verbose_iterable` to make it less public-looking.

pyglmnet/utils.py
```diff
@@ -130,3 +138,41 @@ def set_log_level(verbose):
...
+        Data object wrapped with tqdm.
+    """
+    wrapped_data = data
+    if logger.getEffectiveLevel() == logging.INFO:
+        if has_tqdm:
+            wrapped_data = tqdm(data, leave=False)
+    return wrapped_data
+
+
+def _tqdm_log(msg):
```
geektoni, line 166: @jasmainak I've added this new method here to switch easily from the standard logger to tqdm if we need to print some information within a tqdm loop. This makes sure that everything will look good inside the terminal/notebook.

geektoni:
> Otherwise, LGTM. Can you set the PR to MRG if you're ready?

@jasmainak  I've updated some minor things. If you think it's good I'll move to MRG.

pyglmnet/utils.py
```diff
@@ -130,3 +138,41 @@ def set_log_level(verbose):
             raise ValueError('verbose must be of a valid type')
         verbose = logging_types[verbose]
     logger.setLevel(verbose)
+
+
+def _verbose_iterable(data):
```
jasmainak, line 143: humm ... does this respect the `verbose` argument that is passed to the `GLM` class?

geektoni: Yes, it does. It basically checks the logger's level and if it is appropriate, then it makes the progress interactive.

jasmainak: okay looks good to me then! ready for merge?

geektoni: Yup, I've just changed the PR's title.

jasmainak:
Thanks heaps @geektoni !!

PR #392, opened 2020-08-16
[WIP] Add GLM example with the Negative Binomial distribution.

geektoni:
This example is related to issue #386.

### DIFF
--- a/examples/plot_negative_binomial.py
+++ b/examples/plot_negative_binomial.py
@@ -1,0 +1,75 @@
+# -*- coding: utf-8 -*-
+"""
+=======================================
+GLM with Negative Binomial Distribution
+=======================================
+
+This is an example of GLM with negative binomial distribution.
+We wrote this example taking inspiration from the R community
+below
+https://stats.idre.ucla.edu/r/dae/negative-binomial-regression/
+
+Here, we would like to predict the days absence of high school
+juniors at two schools from there type of program they are enrolled,
+and their math score.
+"""
+
+########################################################
+
+# Author: Titipat Achakulvisut <my.titipat@gmail.com>
+#         Giovanni De Toni <giovanni.det@gmail.com>
+# License: MIT
+
+########################################################
+
+
+########################################################
+# Import relevance libraries
+
+import pandas as pd
+from pyglmnet import GLM
+
+from sklearn.model_selection import train_test_split
+
+import matplotlib.pyplot as plt
+
+########################################################
+# Read and preprocess data
+df = pd.read_stata("https://stats.idre.ucla.edu/stat/stata/dae/nb_data.dta")
+
+# Histogram of type of program they are enrolled
+df.hist(column='daysabs', by=['prog'])
+plt.show()
+
+# Print mean and standard deviation for each program enrolled
+df.groupby('prog').agg({'daysabs': ['mean', 'std']})
+
+########################################################
+# Feature
+X = df.drop('daysabs', axis=1)
+y = df['daysabs'].values
+
+# design matrix
+program_df = pd.get_dummies(df.prog)
+Xdsgn = pd.concat((df['math'], program_df.drop(3.0, axis=1)), axis=1).values
+
+# Split the dataset into training and test
+Xtrain, Xtest, ytrain, ytest = train_test_split(Xdsgn, y, test_size=0.2)
+
+########################################################
+# Fit the model using the GLM
+glm_neg_bino = GLM(distr='neg-binomial',
+                   alpha=0.0,
+                   reg_lambda=0.0,
+                   max_iter=10000,
+                   score_metric='pseudo_R2')
+glm_neg_bino.fit(Xtrain, ytrain)
+
+########################################################
+# Predict
+y_hat = glm_neg_bino.predict(Xtest)
+
+########################################################
+# Return the learned betas and the score
+print(glm_neg_bino.beta0_, glm_neg_bino.beta_)
+print(glm_neg_bino.score(Xtest, y_hat))


jasmainak:
Nice! can you explain somewhere why negative binomial is appropriate for this data?

geektoni:
> Nice! can you explain somewhere why negative binomial is appropriate for this data?

Sure. I'll expand the description of the example.

geektoni:
I think this is clearer now.

examples/plot_negative_binomial.py
```diff
@@ -0,0 +1,94 @@
...
+dispersion.
+"""
+
+########################################################
+
+# Author: Titipat Achakulvisut <my.titipat@gmail.com>
+#         Giovanni De Toni <giovanni.det@gmail.com>
+# License: MIT
+
+########################################################
```
jasmainak, line 39:
```suggestion
```

I can see where this is coming from but I suggest conforming to the other examples :-)

examples/plot_negative_binomial.py
```diff
@@ -0,0 +1,94 @@
...
+
+The Negative Binomial regression can be seen as a mixture of Poisson
+regression in which the mean of the Poisson distribution can be seen
+as a random variable drawn from a Gamma distribution.
+
+This gives us an extra parameter which can be used to account for the over
+dispersion.
+"""
+
+########################################################
```
jasmainak, line 33:
```suggestion
```

jasmainak:
Excellent! I am going to try the code out later this afternoon :-)

examples/plot_negative_binomial.py
geektoni: which examples are you referring to? I used as a model the ones inside `pyglmnet/examples` and they format things the same way :/

jasmainak: oh well, you are right :-) It's different from other repositories I contribute to but I can live with it

jasmainak:
@geektoni I tried it. Something is off. When I plot the convergence:

```py
>>> glm_neg_bino.plot_convergence()
```

I see this plot:

convergence

it should go down monotonically

this could explain why we don't see comparable results

examples/plot_negative_binomial.py
```diff
@@ -0,0 +1,92 @@
...
+# design matrix
+program_df = pd.get_dummies(df.prog)
+Xdsgn = pd.concat((df['math'], program_df.drop(3.0, axis=1)), axis=1).values
+
+# Split the dataset into training and test
+Xtrain, Xtest, ytrain, ytest = train_test_split(Xdsgn, y, test_size=0.2)
+
+########################################################
+# Fit the model using the GLM
+glm_neg_bino = GLM(distr='neg-binomial',
```
jasmainak, line 78:
```suggestion
glm_nb = GLM(distr='neg-binomial',
```

examples/plot_negative_binomial.py
```diff
@@ -0,0 +1,92 @@
...
+# Feature
+X = df.drop('daysabs', axis=1)
+y = df['daysabs'].values
+
+# design matrix
+program_df = pd.get_dummies(df.prog)
+Xdsgn = pd.concat((df['math'], program_df.drop(3.0, axis=1)), axis=1).values
+
+# Split the dataset into training and test
+Xtrain, Xtest, ytrain, ytest = train_test_split(Xdsgn, y, test_size=0.2)
```
jasmainak, line 74: should we avoid doing the training/test split to mirror the R result as closely as possible?

examples/plot_negative_binomial.py
```diff
@@ -0,0 +1,92 @@
...
+from sklearn.model_selection import train_test_split
+
+import matplotlib.pyplot as plt
+
+########################################################
+# Read and preprocess data
+df = pd.read_stata("https://stats.idre.ucla.edu/stat/stata/dae/nb_data.dta")
+
+# Histogram of type of program they are enrolled
+df.hist(column='daysabs', by=['prog'])
```
jasmainak, line 55: possible to add labels to the plots rather than "1", "2", "3"?

pavanramkumar:
> it should go down monotonically
> 
> this could explain why we don't see comparable results

i saw the same issue in convergence plots when i tried to reproduce the R example. had to tweak (increase) learning rates to get it go down monotonically.

pavanramkumar:
showing poisson vs neg bin fits in our example could be useful. i also like this example: https://data.library.virginia.edu/getting-started-with-negative-binomial-regression-modeling/

it talks about where poisson assumptions are lacking and how neg bin could be a better choice.

jasmainak:
> i saw the same issue in convergence plots when i tried to reproduce the R example. had to tweak (increase) learning rates to get it go down monotonically.

that's a bit odd. I can imagine non-monotonic with larger learning rates but I don't understand why it would be the case for smaller learning rates?

jasmainak:
@geektoni I pushed a bunch of fixes. However, I noticed that the link function they use is log which is different from the one we use. So we can't really expect equivalent solutions. I'm wondering if we should wait on #390 to be merged and try with a log link function (should be a couple of lines of code). I think it would be a good validation exercise.

jasmainak:
> showing poisson vs neg bin fits in our example could be useful.

this is a great idea too!

geektoni:
> @geektoni I pushed a bunch of fixes. However, I noticed that the link function they use is log which is different from the one we use. So we can't really expect equivalent solutions. I'm wondering if we should wait on #390 to be merged and try with a log link function (should be a couple of lines of code). I think it would be a good validation exercise.

I guess we could add another example (besides this one) in which we show how to tweak an existing model to use a different link function. However, I remember that, at first, we tried coding the negative binomial with a `log` link function but we had some problems of numerical instability (e.g., nans when computing the gradients). 

I think it is fine if pyglmnet offers a model with a different link function than other libraries. We might not be able to compare everything one-to-one, but as long as the other tests/examples pass I think we can be fairly sure they do the same thing.

geektoni:
> > showing poisson vs neg bin fits in our example could be useful.
> 
> this is a great idea too!

I've added the Poisson regression to the example. As usual, it plots the learned betas and the convergence plot.

examples/plot_negative_binomial.py
geektoni: This is now updated such to show the program names instead (e.g., General, Vocational and Academic).

jasmainak:
@geektoni it might be really helpful to reproduce exactly an example from another source. I tried with log link by rebasing this branch against the `distribution` branch locally. But it didn't change much for me. You could also try the same.

geektoni:
> showing poisson vs neg bin fits in our example could be useful. i also like this example: https://data.library.virginia.edu/getting-started-with-negative-binomial-regression-modeling/
> 
> it talks about where poisson assumptions are lacking and how neg bin could be a better choice.

@jasmainak @pavanramkumar I have added another example taken from the link above. This time it seems to me that we are getting the same results!

jasmainak:
Sounds promising, I'm going to look more closely tomorrow :)

examples/plot_neg_binomial_homicide_victims.py
```diff
@@ -0,0 +1,114 @@
...
+             score_metric='pseudo_R2',
+             verbose=True,
+             max_iter=5000,
+             learning_rate=1e-1)
+glm_poisson.fit(Xdsgn, y)
+print(glm_poisson.beta0_, glm_poisson.beta_)
+
+########################################################
+# Plot convergence information for both negative binomial and poisson
+glm_nb.plot_convergence()
```
jasmainak, line 98: the learning rate seems too high ... this plot really should be monotonically decreasing and reach tol, otherwise you don't get the correct result

examples/plot_neg_binomial_homicide_victims.py
```diff
@@ -0,0 +1,114 @@
...
+# Fit the model using the Poisson regression instead
+glm_poisson = GLM(distr='poisson',
+             alpha=0.0,
+             reg_lambda=0.0,
+             score_metric='pseudo_R2',
+             verbose=True,
+             max_iter=5000,
+             learning_rate=1e-1)
+glm_poisson.fit(Xdsgn, y)
+print(glm_poisson.beta0_, glm_poisson.beta_)
```
jasmainak, line 94: so the coefficient seems to match for Poisson: 1.73 but not for NB. And the intercept is different. Am I missing something?

examples/plot_neg_binomial_homicide_victims.py
```diff
@@ -0,0 +1,114 @@
...
+#
+# The Poisson model outputs the predicted mean (and therefore variance) of the distribution.
+# However, we can see from the exploratory analysis that the observed standard deviations are much larger.
+# The Negative Binomal generates the same mean, but we can use the dispersion parameter to
+# compute a more accurate estimate of the standard deviations (for both white and black classes).
+#
+pred_nb = np.array(glm_nb.predict([[0], [1]]))
+pred_poisson =  np.array(glm_poisson.predict([[0], [1]]))
+print("")
+print("NB Predicted means+std (black/white): {}, {}".format(pred_nb, np.sqrt(pred_nb+(pred_nb**2)*1/0.20)))
```
jasmainak, line 113: not sure I follow where this expression comes from but I guess we can clarify this later. First step is to make sure we get the same results

jasmainak (commented):
@geektoni sorry for taking so long, it has been a busy week. I have a couple of questions. If required, you and me can do a quick skype in the coming week to push this forward. I would really love to see this in the repo!

